PR #4, opened 2017-01-23
Create structure for fraction module within commons-numbers.

RayDeCampo:
Add dependencyManagement to parent pom.xml for uniform dependency on numbers core; remove version from references to numbers-core in child modules.
Add test jar to numbers core pom.xml to allow for re-use of TestUtils.
Add ArithmeticUtils from commons-math since Fraction, BigFraction heavily depends on it.

I pulled the classes from the develop branch of commons-math.
I had to do something with the exceptions, I created a handful but I think we will want to revisit and perhaps make them more cohesive.
As noted above I pulled in ArithmeticUtils from commons-math.
I did not make a lot of changes to the comments, they probably need to be looked over.


RayDeCampo:
Forgot to mention that I commented out the references to the Field and FieldElement interfaces (but left the methods that implement them).  Not sure if the plan was to bring those into numbers core.

PR #20, opened 2018-12-14
NUMBERS-84: Checks for NaN and Infinite in Quaternion Normalization

darkma773r:
Adding NaN and infinite checks to Quaternion.normalize() and Quaternion.inverse(). I ended up bumping the JDK version from 6 to 8 in order to use Double.isFinite(). If this causes issues, I can rework it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ language: java
 sudo: false
 
 jdk:
-  - openjdk7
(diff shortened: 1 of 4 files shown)


darkma773r:
I'm closing this since master has changed and it will be easier to open a new one than try to resolve the conflicts.

PR #26, opened 2018-12-23
[NUMBERS-88] Don't use the default locale in message formatting

leventov:
https://issues.apache.org/jira/browse/NUMBERS-88

This patch introduces forbidden-apis checker to catch violations in the future.


aherbert:
Out of date and requires large refactoring. Message formatting is no longer used.

PR #28, opened 2019-01-20
NUMBERS-90: adding DoublePrecisionContext and FloatPrecisionContext

darkma773r:
Adding DoublePrecisionContext and FloatPrecisionContext abstract classes along with simple epsilon-based implementations. I also created a package named "precision" for these and moved the "Precision" utility class there as well.

These changes are required for GEOMETRY-11.


darkma773r:
I'm closing this PR since this functionality was placed directly in commons-geometry, at least until the API and use cases can be explored more. See GEOMETRY-11.

PR #31, opened 2019-02-24
Fix missing negation in Complex#sin()

renggli:
The comment message of the method is actually correct, but the implementation lacks the negation.


aherbert:
Complex has changed significantly since this PR was made. Please open another PR if you find any more issues.

PR #41, opened 2019-05-30
[NUMBERS-107] - Cleanup build configuration - remove not used configu…

khmarbaise:
…rations


aherbert:
I don't think we need this PR. As soon as numbers is released (a beta1 is currently out now) then we will have to add back the text about binary compatibility.

aherbert:
Closing as not required.

PR #47, opened 2019-06-05
[numbers] support JUnit 5 tests

grimreaper:

PR #50, opened 2019-06-06
NUMBERS-113: Fix JUnit dependencies in pom.xml

Schamschi:
I'm not a JUnit expert, so I'm not sure if this is the optimal way to import both JUnit 5 and JUnit 4. But it does seem to solve the problems described in the JIRA ticket, so I'll just be bold and put this here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,12 +103,19 @@
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-engine</artifactId>
-      <version>5.0.0-M4</version>
+      <version>5.4.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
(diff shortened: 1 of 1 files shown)


aherbert:
I think that @grimreaper may have pushed over the original PR #47 that passed all the checks with a different revision. Hopefully it can be sorted quickly.

grimreaper:
hrm, you're right. I thought I pushed the version with the BOM but apparently not.

grimreaper:
@aherbert I think I fixed it now.

grimreaper:
See 6bcc6d9a16bea265243f68b298713d10abb09224

aherbert:
The JUnit change is now restored in master. @Schamschi can you confirm if this has fixed your JUnit issues on your branch and close?

Schamschi:
@aherbert Yes, it seems fixed now, I'm going to close the pull request and the JIRA ticket.

PR #56, opened 2019-06-22
NUMBERS-119: Correct bug in BigFraction(double) constructor

Schamschi:
Take care of special exponent bias with subnormal numbers


Schamschi:
Interesting, the code coverage decreased, but not because of the special case I added for subnormal numbers (this was already covered by a test that only tested the numerator but not the denominator, which is why the test did not fail), but because I also made a special case for zero that caused the denominator of the created fraction to be set to 1 (so that the fraction is 0/1) instead of 2^1075. For some reason, this caused a branch in the method reduce() not to be executed by tests anymore …

Schamschi:
I just noticed that the coveralls report is lying about the extent of the decrease in coverage. All the lines in ArithmeticUtils purportedly uncovered by this pull request are already not covered in the master branch (some of them even seem to be unreachable).

PR #58, opened 2019-06-26
NUMBERS-121: Replace usages of deprecated HTML tag <tt> with MathJax

Schamschi:


aherbert:
No reason to close this. The deprecated `<tt>` tag should be replaced with something. At least then the Javadoc will build. The MathJax can always be reverted to HTML <code> tags later if required. Without this PR the build is broken.

Schamschi:
@aherbert Yeah, you're right, my closing of this pull request was a bit rash.

PR #66, opened 2019-07-18
[NUMBERS-131] Re-write implementations of double approximation factory methods in BigFraction

Schamschi:


commons-numbers-fraction/src/main/java/org/apache/commons/numbers/fraction/BigFraction.java
```diff
@@ -320,14 +320,103 @@ public static BigFraction from(final double value,
...
     public static BigFraction from(final double value,
-                                   final int maxDenominator) {
-        return from(value, 0, maxDenominator, 100);
+                                   final BigInteger maxDenominator) {
+        if (maxDenominator.signum() != 1) {
+            throw new IllegalArgumentException("Upper bound for denominator must be positive: " + maxDenominator);
+        }
+
+        /*
+         * Required facts:
```
ericbarnhill: At some point it will probably be better to move discussions like this to a wiki or other documentation, and just link to it here.

ericbarnhill:
Thanks for the great contribution Heinrich! I can't figure out why it's giving you a conflict, looks like you just deleted that code to me :).

Schamschi:
Hi Eric,

sorry for the late reply, I have a bit much to do right now, and I don't think I'll be able to have a look at it before Sunday. But thanks for taking the time to look at it. There are some other things that need to be changed besides those you mentioned, for example the Javadoc, as I've learned in the meantime that MathJax is preferred over HTML also in private and package-private elements.

And about the lengthy explanation of the algorithm: Do you have a specific suggestion? The essence of the algorithm is described in the Wikipedia article Continued Fraction, which is where I got it from, but I'm not sure if this is a suitable source to cite, because it doesn't contain any explanation as to why the algorithm works, and the references for the respective sections of the article were not very helpful either, as far as I remember. The book by A. Khinchin was the best source I could find, and the sections I cited are available on the free preview on Google Books, but I couldn't find a summary of the algorithm there (at least not in the free preview, I don't have the book, so I don't know if there is one somewhere in the rest of the book).

aherbert:
This PR is out-of-date. The algorithm for BigFraction was rewritten since this PR was made. But IIUC this contains a method to extend beyond the `long` datatype for the numerator/denominator used in the current implementation.

If you care to rebase the code on the current master then I'll take a look.

PR #68, opened 2019-07-24
[NUMBERS-133] Use iteration algorithm from bounded trial division in Primes.nextPrime(int)

Schamschi:
I'm not sure if "potentialPrimes" is a good name for the method that generates the iterator, because it doesn't reveal anything about the purpose of the argument (a lower bound). However, "potentialPrimesBeyond" or something similar would be wrong, because the argument itself is also included in the possible range of results. The most descriptive name would be "potentialPrimesGreaterThanOrEqualTo", but this is horribly verbose. "potentialPrimesNotSmallerThan" is only slightly less verbose, and it contains a negation, which is confusing. So I've left it at "potentialPrimes", but any better suggestions are welcome. At least it's a package-private method, so the method name could still be changed any time in the future.

### DIFF
--- a/commons-numbers-primes/src/main/java/org/apache/commons/numbers/primes/Primes.java
+++ b/commons-numbers-primes/src/main/java/org/apache/commons/numbers/primes/Primes.java
@@ -17,7 +17,10 @@
 package org.apache.commons.numbers.primes;
 
 import java.text.MessageFormat;
+
+import java.util.Arrays;
 import java.util.List;
+import java.util.PrimitiveIterator;
 
 
 /**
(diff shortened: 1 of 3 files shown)


aherbert:
How about `potentialPrimesGTE`? The javadoc can just contain (GTE) in brackets after the `greater than or equal to` text.

[GTE](https://www.acronymfinder.com/Greater-Than-or-Equal-(mathematics)-(GTE).html)

Schamschi:
@aherbert That sounds like I good idea. I had also thought about using an abbreviation, but I only considered the one from the LaTeX tag "geq", which would probably have been too LaTeX specific. I don't know how common "gte" is, but chances are it will still prevent more confusion than it causes. Thanks.

aherbert:
@Schamschi  The merge of this has been missed. Would you care to rebase this on the current master?

PR #69, opened 2019-10-23
Big fraction unit tests

bkrogh:
Commit 44ea3e26fa7610f3635aa225d24c4573ad5e54f2 breaks BigFraction so that division does not work anymore.
The snapshot release of this repository is used by other apache projects, e.g., apache commons geometry, which are currently failing to build too.


bkrogh:
I confirm the that commit 2f5a658 fixes the reported issue. :)

Thanks.

PR #71, opened 2019-12-04
NUMBERS-136: log10() uses log() functionality

ericbarnhill:
log10() greatly simplified and uses log() functionality


aherbert:
Hi @ericbarnhill, I think you have based this on an old version of master.

I fixed log10 when fixing the ISO C99 reference test. I changed the implementation to use Math.log10 in place of Math.log.

Have a look at the current master.

ericbarnhill:
You are right, I thought I pulled in the latest master before I began work, but somehow did not

PR #75, opened 2020-02-13
NUMBERS-144: Fixing rat check failures from examples module

darkma773r:
Adding standard rat exclusions to prevent failures from files in the examples module.


darkma773r:
This issue was fixed in other commits.

PR #84, opened 2020-08-25
[NUMBERS-151] performance refine for ArithmeticUtils.pow

XenoAmess:


aherbert:
You have added no tests. There are a few lines not covered by the existing tests.

XenoAmess:
@aherbert 
> You have added no tests. There are a few lines not covered by the existing tests.

I will go see the coverage and add tests.

BTW.
Should we really throw no exception when we invoke pow(0, 0) here?
It seems...weird...

XenoAmess:
@aherbert 
tests added.

aherbert:
`0^0` is undefined. Here we support returning 1 as described in this Wikipedia article:
[Zero_to_the_power_of_zero](https://en.wikipedia.org/wiki/Zero_to_the_power_of_zero)

Perhaps add a comment:
```
<p>Special cases:
<ul>
<li>{@code k^0} returns {@code 1} (including {@code k=0})
<li>{@code k^1} returns {@code k} (including {@code k=0})
<li>{@code 0^0} returns {@code 1}
<li>{@code 0^e} returns {@code 0}
<li>{@code 1^e} returns {@code 1}
<li>{@code (-1)^e} returns {@code -1 or 1} if {@code e} is odd or even
</ul>
```

commons-numbers-core/src/test/java/org/apache/commons/numbers/core/ArithmeticUtilsTest.java
```diff
@@ -435,6 +435,25 @@ void testIsPowerOfTwo() {
         }
     }
 
+    @Test
+    void testPowOptimize() {
```
aherbert, line 439: This should be moved above the `testIsPowerOfTwo` method. I would rename it to `testPowEdgeCases`.

aherbert (changes requested):
A final fix for the new tests and then can you squash all the commits to a single commit.

commons-numbers-core/src/main/java/org/apache/commons/numbers/core/ArithmeticUtils.java
```diff
@@ -266,6 +266,16 @@ public static long lcm(long a, long b) {
     /**
      * Raise an int to an int power.
      *
+     * <p>Special cases:</p>
```
aherbert, line 269: You should put the same comment on the `long` version of the pow method.

XenoAmess: @aherbert Ah, I forgot it :)
will do now.

commons-numbers-core/src/test/java/org/apache/commons/numbers/core/ArithmeticUtilsTest.java
XenoAmess: @aherbert done.

PR #87, opened 2020-12-24
Minor Improvement:

arturobernalg:
* Add final
* Unnecessary semicolon
* Fix JavaDoc
* Remove Unnecessary semicolon

### DIFF
--- a/commons-numbers-angle/src/main/java/org/apache/commons/numbers/angle/PlaneAngle.java
+++ b/commons-numbers-angle/src/main/java/org/apache/commons/numbers/angle/PlaneAngle.java
@@ -40,7 +40,7 @@ public final class PlaneAngle {
     /**
      * @param value Value in turns.
      */
-    private PlaneAngle(double value) {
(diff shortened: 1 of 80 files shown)


aherbert:
The 'final' status of parameters is enforced in the project using PMD rules: `AvoidReassigningParameters`. However we have avoided using final in the code as it adds visual noise to the source code and the PMD rule `MethodArgumentCouldBeFinal` is disabled. There is no benefit of final method arguments to the JVM compiler, which knows about effectively final variables for example those that can be referenced inside lambda statements.

As such it is difficult to see what other modifications are in the PR. Can you remove the use of final for method arguments and then it will be clear what is left?

The use of final for method params is a style choice and you can raise this on the dev mailing list to start a discussion about what is preferred.

garydgregory:
This is my pitch for using final: https://www.google.com/amp/s/garygregory.wordpress.com/2013/01/26/the-final-kiss-in-java/

PR #90, opened 2021-04-23
NUMBERS-142: Update LinearCombination to use the dot2s algorithm

aherbert:
- Avoids construction of an intermediate array for the dot product of
array input
- Perform splitting using Dekker's multiplication algorithm to retain
all bits of precision
- Handle overflow during the split for large numbers

Splitting now correctly handles sub-normal numbers with no information
in the upper 26-bits as input. The high part will be sub-normal and the
low part will be zero. The previous split created a zero high part and
the input as the low part.


darkma773r (approved):
Looks good to me!

PR #92, opened 2021-05-28
NUMBERS-156: replacing SafeNorm with Norms and Summation

darkma773r:
- replacing SafeNorm with Norms
- adding Summation class
- adding norm algorithm evaluation tests and benchmarks

Note that this change requires JDK 9+ for examples-jmh and commons-numbers-arrays unit tests.


commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/Norms.java
```diff
@@ -0,0 +1,427 @@
...
+            return Double.POSITIVE_INFINITY;
+        }
+
+        // compute the scale and rescale values
+        final double scale;
+        final double rescale;
+        if (max > LARGE_THRESH) {
+            scale = SCALE_DOWN;
+            rescale = SCALE_UP;
+        } else if (max < SAFE_SCALE_UP_THRESH) {
```
darkma773r, line 153: @aherbert, this is slightly different than the implementation in the ExtendedPrecisionLinearCombination algorithm in examples-jmh. In that class, we scale up if the max is under a threshold for safely scaling up _and_ the min value is less than the threshold for very small numbers. My thought here is that we only need to check the max. If the max value is small enough that we can scale up without overflow, then we may as well just go ahead and do it and not bother checking (or even computing) the min. Let me know if you agree with this assessment.

aherbert: The logic makes sense since you only have two numbers. Note you can change this to a new value:
```java
SAFE_SCALE_UP_THRESH = 0x1.0p-100;
```
You are scaling by 2^600. So this value scaled would be 2^500 and safe against overflow when squared.

With two numbers you do not need to do the computation if the difference between them is large because max^2 + min^2 is mute:
```java
if (max - min > 0x1.0p54) {
    return max;
}
```
So knowing min is useful for a fast exit.

I'll do a full review of this later. But the thresholds should be changed to 2^496 from 2^500 for the array case.

darkma773r: Why the change for the array case? Also, should/could that apply to the other cases as well?

aherbert: Because if you have a lot of memory you can pass in an array of 2^31-1 doubles. If these all have a value of Double.MAX_VALUE then you will overflow during the sum unless you lower the threshold. When you only have 2 or 3 doubles then you could make the threshold higher.

I think you need 2^496 for the array case and 2^510 for the case of 3 inputs. This would be 2^992 or 2^1020 when squared. You can add respectively 2^31 of the former and 3 of the latter.

aherbert: I should add that due to the extended precision summation you do not want it to sum to infinite. This will lead to errors in the round-off computation that creates NaN. So even if the final result would be infinite you do not want it to return NaN. So the thresholds should allow the summation to complete to a finite value. The final scaling can then create the infinite if the result is not possible within the range of a double.

darkma773r: I've updated the threshold and added the early exit check for Euclidean 2D. I don't see a compelling reason to use two separate thresholds, so all Euclidean methods use the 2^496 threshold. The early exit in the 2D case gave a small improvement in performance in the JMH benchmark, small enough where the outliers of the before and after case overlap.

darkma773r: Let me know if you see any other changes that should be made.

darkma773r: @aherbert, did you still want to review this or do you think it's ready to go?

aherbert: Sorry, I have been busy. I will try and find time this weekend.

darkma773r: No worries! I was just making sure there wasn't something I was missing here.

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/Norms.java
```diff
@@ -0,0 +1,451 @@
...
+     * @see <a href="https://en.wikipedia.org/wiki/Norm_(mathematics)#Maximum_norm_(special_case_of:_infinity_norm,_uniform_norm,_or_supremum_norm)">Maximum norm</a>
+     */
+    public static double maximum(final double x, final double y, final double z) {
+        return Math.max(
+                Math.abs(x),
+                Math.max(Math.abs(y), Math.abs(z)));
+    }
+
+    /** Compute the maximum norm (also known as the infinity norm or L<sub>inf</sub> norm) of the given values.
+     * The result is equal to \(\max{(|v_0|, ... |v_i|)}\), i.e., the maximum of the absolute values of the
```
aherbert, line 431: This should be `\(\max{(|v_0| + ... + |v_i|)}\)` to match the style of the other javadocs.

I rendered the javadocs and prefer the `\ldots`. Also use the length of the input n so the summation is to n-1.

Edit to `\(\max{(|v_0| +  \ldots +  |v_{n-1}|)}\)`.

```
rm -f target/maven-javadoc-plugin-stale-data.txt
mvn javadoc:javadoc
open target/site/apidocs/org/apache/commons/numbers/arrays/Norms.html
```

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/DoubleTestUtils.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param maxExp maximum exponent
+     * @param rng random number generator
+     * @return random double value with an exponent in the specified range
+     */
+    public static double randomDouble(final int minExp, final int maxExp, final UniformRandomProvider rng) {
+        // Create random doubles using random bits in the sign bit and the mantissa.
+        final long mask = ((1L << 52) - 1) | 1L << 63;
+        final long bits = rng.nextLong() & mask;
+        // The exponent must be unsigned so + 1023 to the signed exponent
+        final int expRange = Math.abs(maxExp - minExp);
```
aherbert, line 66: Should this be Math.abs(maxExp - minExp) + 1?

Otherwise the call to rng.nextInt(int) could be called with 0 when max is the same as min.

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/ExtendedPrecisionTest.java
```diff
@@ -102,6 +102,23 @@ void testSubNormalSplit() {
         final double lo2 = a - hi2;
         Assertions.assertEquals(a, hi2);
         Assertions.assertEquals(0, lo2);
+
```
aherbert, line 105: Remove extra line

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/ExtendedPrecisionTest.java
```diff
@@ -102,6 +102,23 @@ void testSubNormalSplit() {
...
+
         Assertions.assertTrue(Math.abs(hi2) > Math.abs(lo2));
     }
+
+    @Test
+    void testSquareLowUnscaled() {
+        assertSquareLowUnscaled(0.0, 1.0);
+        assertSquareLowUnscaled(0.0, -1.0);
+        assertSquareLowUnscaled(-1.4293872661474477E-16, Math.PI);
```
aherbert, line 114: Can you compute this result using BigDecimal?
```java
final BigDecimal pi = new BigDecimal(Math.PI);
final double pi2low = pi.pow(2).subtract(new BigDecimal(Math.PI * Math.PI)).doubleValue();
assertSquareLowUnscaled(pi2low, Math.PI);
```

It is a bit cryptic to have a constant here.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/DoubleUtils.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param maxExp maximum exponent value
+     * @param rng random number generator
+     * @return random double
+     */
+    public static double random(final int minExp, final int maxExp, final UniformRandomProvider rng) {
+        // Create random doubles using random bits in the sign bit and the mantissa.
+        final long mask = ((1L << 52) - 1) | 1L << 63;
+        final long bits = rng.nextLong() & mask;
+        // The exponent must be unsigned so + 1023 to the signed exponent
+        final long exp = rng.nextInt(Math.abs(maxExp - minExp)) + minExp + 1023;
```
aherbert, line 39: Again the range should have a +1 to avoid calling nextInt with the wrong limit.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/DoubleUtils.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.numbers.examples.jmh;
+
+import org.apache.commons.rng.UniformRandomProvider;
+
+/** Class containing utility methods for working with doubles.
+ */
+public final class DoubleUtils {
```
aherbert, line 23: Does all this have to be public? I think it can move to the arrays package and be package private.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/EuclideanNormAlgorithmPerformance.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /** String indicating double exponents with mid-level values which will not overflow or underflow. */
+    private static final String MID = "mid";
+
+    /** String indicating double exponents with very high positive values, likely to overflow. */
+    private static final String HIGH = "high";
+
+    /** String indicating double exponents over a very wide range of values. */
+    private static final String FULL = "full";
+
```
aherbert, line 66: Remove the extra line

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/EuclideanNormAlgorithmPerformance.java
```diff
@@ -0,0 +1,235 @@
...
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+@Fork(value = 1, jvmArgs = {"-server", "-Xms512M", "-Xmx512M"})
+public class EuclideanNormAlgorithmPerformance {
+
+    /** Default number of samples used in each benchmark. */
+    private static final int DEFAULT_SAMPLES = 100_000;
```
aherbert, line 50: This is only used in one place. Just hard code it there and avoid the ugly `int + ""` declaration in the `@Param` tag.

Same for DEFAULT_VECTOR_LEN.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/EuclideanNormAlgorithmPerformance.java
```diff
@@ -0,0 +1,235 @@
...
+                maxExp = +10;
+                break;
+            case HIGH:
+                minExp = +510;
+                maxExp = +530;
+                break;
+            default:
+                minExp = -550;
+                maxExp = +550;
+                break;
```
aherbert, line 120: I prefer here to throw in the default case statement. So if you pass parameters to JMH on the command line it will fail:
```
mvn package -Pexamples-jmh
java -jar target/examples-jmh.jar EuclideanNormAlgorithmPerformance -p vectorLength=10 -p type=foobar
```

This should error but instead runs with `type=FULL`.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/EuclideanNormEvaluator.java
```diff
@@ -0,0 +1,248 @@
...
+        BigDecimal sum = BigDecimal.ZERO;
+        for (final double v : vec) {
+            sum = sum.add(BigDecimal.valueOf(v).pow(2), ctx);
+        }
+
+        return sum.sqrt(ctx).doubleValue();
+    }
+
+    /** Compute the ulp difference between two values. This method assumes that
+     * the arguments have the same exponent.
```
aherbert, line 106: They do not have to have the same exponent. The ulp difference between 1.0 and Math.nextDown(1.0) will be 1. They do have to have the same sign. All your results should be positive so this is OK. 

If the numbers are very different then this will return a truncated difference by long to int primitive conversion. A return type of `long` would fix this. I do not think that you require `int` during testing as the sums use a double.

This comment also applies to `DoubleTestUtils` in the main package.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/EuclideanNormEvaluator.java
```diff
@@ -0,0 +1,248 @@
...
+            double diffSumSq = 0d;
+            double diff;
+            for (double ulpError : ulpErrors) {
+                if (Double.isFinite(ulpError)) {
+                    diff = ulpError - mean;
+                    diffSumSq += diff * diff;
+                }
+            }
+
+            final double stdDev = Math.sqrt(diffSumSq / (successCount - 1));
```
aherbert, line 243: In the ridiculous case where success count is 1 this will be invalid. zero successes will compute a NaN mean which is a useful marker, and stdDev will be zero.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/NormsPerformance.java
```diff
@@ -0,0 +1,194 @@
...
+         */
+        @Setup
+        public void createVectors() {
+            final UniformRandomProvider rng = RandomSource.create(RandomSource.XO_RO_SHI_RO_1024_PP);
+
+            vectors = new double[SAMPLES][];
+            for (int i = 0; i < vectors.length; ++i) {
+                // pick a general range for the vector element exponents and then
+                // create values within that range
+                final int vMidExp = rng.nextInt(MAX_EXP - MIN_EXP) + MIN_EXP;
```
aherbert, line 86: Invalid range again. Add a 1 to it.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/NormsPerformance.java
```diff
@@ -0,0 +1,194 @@
...
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+@Fork(value = 1, jvmArgs = {"-server", "-Xms512M", "-Xmx512M"})
+public class NormsPerformance {
+
+    /** Number of samples used in each benchmark. */
+    private static final int SAMPLES = 100_000;
```
aherbert, line 50: Can all these be put into `@Param` annotated attributes in VectorArrayInput. This allows them to be adjusted by passing values to JMH on the command line.

commons-numbers-examples/examples-jmh/src/test/java/org/apache/commons/numbers/examples/jmh/arrays/EuclideanNormAccuracyTest.java
```diff
@@ -0,0 +1,142 @@
...
+                    stats.getUlpErrorMean(),
+                    stats.getUlpErrorStdDev(),
+                    stats.getUlpErrorMin(),
+                    stats.getUlpErrorMax(),
+                    stats.getFailCount()));
+            writer.newLine();
+        }
+    }
+
+    /** Create an array of random input vectors with exponents in the range {@code [minExp, maxExp]}..
```
aherbert, line 116: Two periods at the end of the line.

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/NormsTest.java
```diff
@@ -0,0 +1,564 @@
...
+        for (int i = 0; i < samples; ++i) {
+            exactResults[i] = exactEuclideanNorm(inputs[i]);
+        }
+
+        // compute the std devs
+        final double hypotResult = computeUlpErrorStdDev(inputs, exactResults, v -> Math.hypot(v[0], v[1]));
+        final double normResult = computeUlpErrorStdDev(inputs, exactResults, v -> Norms.euclidean(v[0], v[1]));
+
+        // ensure that we are within the ballpark of Math.hypot
+        Assertions.assertTrue(normResult <= (hypotResult + HYPOT_COMPARE_EPS),
```
aherbert, line 222: You should also check the mean. What if Norms computed the exact result but shifted by 20 ULP. It would have a standard deviation of zero but a mean of 20. It would be very precise in its inaccuracy. Effectively you should test accuracy (mean ulp) and precision (std dev of the ulp).

aherbert (changes requested):
The code looks good. The `Norms.euclidean(double, double)` should be fixed to correctly detect NaN as intended, not by computing and then returning NaN. I have added a few notes on minor nits which you can fix as you see fit.

aherbert:
One further note is that this change seems to break coveralls reporting. The current master is working fine. You can see the coveralls error in the Travis log. I am not sure why but this can be investigated separately. It may be due to the use of JDK 9 for travis. Perhaps bump it to JDK 11.

aherbert:
The problems with coveralls has been reported elsewhere for JDK 9, e.g:
[coveralls-maven-plugin fails on Java 9](https://github.com/trautonen/coveralls-maven-plugin/issues/112)

This issue is now closed via a merge so perhaps update the coveralls plugin version in the POM to the latest and see if we get a coverage report.

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/Norms.java
```diff
@@ -0,0 +1,451 @@
...
+     * @param z third input
+     * @return Euclidean norm
+     * @see <a href="https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm">Euclidean norm</a>
+     */
+    public static double euclidean(final double x, final double y, final double z) {
+        final double xabs = Math.abs(x);
+        final double yabs = Math.abs(y);
+        final double zabs = Math.abs(z);
+
+        final double max = Math.max(Math.max(xabs, yabs), zabs);
```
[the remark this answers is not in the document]
darkma773r: The absolute values are used if I use the above fiddle of returning `xabs * yabs * zabs` as mentioned above, which I quite like.

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/Norms.java
darkma773r: I can't use '+' since that would produce the manhattan norm. I'll update the subscript and ellipsis, though.

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/ExtendedPrecisionTest.java
darkma773r: That sounds good but the expected value computed with `BigDecimal` is `-8.79470486217151E-16` whereas the `ExtendedPrecision` result is '-1.4293872661474477E-16'. Is this simply a limitation of the algorithm or is this a bug in our implementation?

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/Norms.java
```diff
@@ -0,0 +1,451 @@
...
+            max = xabs;
+            min = yabs;
+        } else {
+            max = yabs;
+            min = xabs;
+        }
+
+        // if the max is not finite, then one of the inputs must not have
+        // been finite
+        if (!Double.isFinite(max)) {
```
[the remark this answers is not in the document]
darkma773r: I went with the simple version.

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/Norms.java
```diff
@@ -0,0 +1,451 @@
...
+        final double xabs = Math.abs(x);
+        final double yabs = Math.abs(y);
+        final double zabs = Math.abs(z);
+
+        final double max = Math.max(Math.max(xabs, yabs), zabs);
+
+        // if the max is not finite, then one of the inputs must not have
+        // been finite
+        if (!Double.isFinite(max)) {
+            if (Double.isNaN(x) || Double.isNaN(y) || Double.isNaN(z)) {
```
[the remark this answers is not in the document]
darkma773r: I like it!

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/DoubleTestUtils.java
darkma773r: Yep. Oops.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/DoubleUtils.java
darkma773r: Yep. Done.

darkma773r:
@aherbert, thanks for the thorough review! I've implemented the changes you requested except as noted in the comments. Coveralls is still failing, even on JDK 11. There does not appear to be a later version to upgrade to, unfortunately. Let me know if any other changes need to be made.

Note that I also added an exception to the spotbugs exclude file for the new floating point equality expression in Angle.Normalization. The build was failing otherwise.

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/ExtendedPrecisionTest.java
aherbert: I thought the BigDecimal value was the same. I just tried the code above on a different machine and it still works. Stepping through a debugger the value of `pi2low` is `-1.429...`. Are you using a rounding mode for BigDecimal?

aherbert: Since the code is on Java 9 for testing you can drop BigDecimal and simplify this to:
```java
assertSquareLowUnscaled(Math.fma(Math.PI, Math.PI, -Math.PI * Math.PI), Math.PI);
```

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/ExtendedPrecision.java
```diff
@@ -204,6 +204,28 @@ private static double productLowUnscaled(double x, double y, double xy) {
...
+     * Compute the low part of the double length number {@code (z,zz)} for the exact
+     * square of {@code x} using Dekker's mult12 algorithm. The standard precision product
+     * {@code x*x} must be provided. The number {@code x} is split into high and low parts
+     * using Dekker's algorithm.
+     *
+     * <p>Warning: This method does not perform scaling in Dekker's split and large
+     * finite numbers can create NaN results.
+     *
+     * @param x Number to square
+     * @param xx Standard precision product {@code x*x)
```
aherbert, line 217: Javadoc issue: This should have a `}` not a `)`.

commons-numbers-arrays/src/main/java/org/apache/commons/numbers/arrays/ExtendedPrecision.java
```diff
@@ -204,6 +204,28 @@ private static double productLowUnscaled(double x, double y, double xy) {
...
+     * @return the low part of the square double length number
+     * @see #productLowUnscaled(double, double, double)
```
aherbert, line 219: Javadoc issue: The productLowUnscaled is private. So you cannot refer to it in a package private method.

aherbert (commented):
Without coveralls I ran jacoco locally and found 100% coverage. This is good. My IDE picked up a javadoc error in a new method in ExtendedPrecision.

Could you try this fix for coveralls. Add this to the coveralls configuration in the travis profile:
```xml
<version>${commons.coveralls.version}</version>
<!-- Workaround for coveralls plugin JDK 9 compatibility issue.
     https://github.com/trautonen/coveralls-maven-plugin/issues/112 -->
<dependencies>
    <dependency>
        <groupId>javax.xml.bind</groupId>
        <artifactId>jaxb-api</artifactId>
        <version>2.2.3</version>
    </dependency>
</dependencies> 
```

commons-numbers-arrays/src/test/java/org/apache/commons/numbers/arrays/ExtendedPrecisionTest.java
darkma773r: I found why `BigDecimal` wasn't working for me: I was using `BigDecimal.valueOf()` to create instances instead of `new BigDecimal()`. Switching to the constructor to use the exact double representation fixed the issue. I'm switching to the `fma` version, however since that is simpler.

commons-numbers-examples/examples-jmh/src/main/java/org/apache/commons/numbers/examples/jmh/arrays/DoubleUtils.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param maxExp maximum exponent value
+     * @param rng random number generator
+     * @return random double
+     */
+    static double random(final int minExp, final int maxExp, final UniformRandomProvider rng) {
+        // Create random doubles using random bits in the sign bit and the mantissa.
+        final long mask = ((1L << 52) - 1) | 1L << 63;
+        final long bits = rng.nextLong() & mask;
+        // The exponent must be unsigned so + 1023 to the signed exponent
+        final long exp = rng.nextInt(Math.abs(maxExp - minExp) + 1) + minExp + 1023;
```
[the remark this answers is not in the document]
darkma773r: I just documented the requirement.

darkma773r:
The build is passing and it looks like coveralls is working as well, although it's not showing up here in the PR for some reason. Any more changes or do you think we're good to go?

aherbert:
I see that the latest travis build did create a [coveralls report](https://coveralls.io/jobs/81536135). So perhaps it just takes a while to filter through and e-mail the results.

I think this is OK to merge. If you do that make sure to update the change.xml.

Thanks for the effort.

darkma773r:
Thanks! I'll try to merge this tonight and update the changes.xml.

PR #96, opened 2021-06-16
Replace Summation and LinearCombination with Sum

darkma773r:


darkma773r:
Replaced with https://github.com/apache/commons-numbers/pull/97

PR #99, opened 2021-06-20
Replace manual copying of array contents by calls to System.arraycopy().

arturobernalg:


commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/LogFactorial.java
```diff
@@ -62,9 +62,7 @@ private LogFactorial(int numValues,
         }
 
         // Copy available values.
-        for (int i = beginCopy; i < endCopy; i++) {
-            logFactorials[i] = cache[i];
-        }
+        if (endCopy - 2 >= 0) System.arraycopy(cache, beginCopy, logFactorials, beginCopy, endCopy - 2);
```
aherbert, line 65: This will fail checkstyle. The logic seems fine.

commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/FactorialDouble.java
```diff
@@ -64,9 +64,7 @@ private FactorialDouble(int numValues,
         }
 
         // Copy available values.
-        for (int i = beginCopy; i < endCopy; i++) {
-            factorialsDouble[i] = cache[i];
-        }
+        if (endCopy - 2 >= 0) System.arraycopy(cache, beginCopy, factorialsDouble, beginCopy, endCopy - 2);
```
aherbert, line 67: This will fail checkstyle. The logic seems fine.

commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/LogFactorial.java
arturobernalg: True. Changed. TY

commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/FactorialDouble.java
arturobernalg: True. Changed. TY

commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/FactorialDouble.java
```diff
@@ -64,8 +64,8 @@ private FactorialDouble(int numValues,
         }
 
         // Copy available values.
-        for (int i = beginCopy; i < endCopy; i++) {
-            factorialsDouble[i] = cache[i];
+        if (endCopy - 2 >= 0) {
```
aherbert, line 67: There are now null pointer fails in the travis build. You cannot assume `endCopy - 2 >= 0`. Try changing this to `endCopy - beginCopy >= 0`.

arturobernalg: I think same result

aherbert:
If the array is null it is not the same result. Check the logic a few lines
above. Please run the default maven goal before opening PRs so you can fix
the build errors.

PR #103, opened 2021-08-07
Simplify boolean expressions with an simpler and equivalent.

arturobernalg:


commons-numbers-core/src/main/java/org/apache/commons/numbers/core/Precision.java
```diff
@@ -182,7 +182,7 @@ public static boolean equalsIncludingNaN(float x, float y) {
         final boolean yIsNan = Float.isNaN(y);
         // Combine the booleans with bitwise OR
         return (xIsNan | yIsNan) ?
-            !(xIsNan ^ yIsNan) :
+                xIsNan == yIsNan :
```
aherbert, line 185: The formatting looks wrong here as you have 8 character indent and then a 4 character indent on the line underneath. It should be:
```java
        return (xIsNan | yIsNan) ?
            xIsNan == yIsNan :
            equals(x, y, maxUlps);
```

PR #104, opened 2021-08-08
Simplify boolean expressions with an simpler and equivalent.

arturobernalg:
* Simplify boolean expression with equivalent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ bin
 **/target/**
 /build
 /lib
-site-content
+site-content*
 **/.ekstazi/
 *.class
 *.iml


commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/FactorialDouble.java
```diff
@@ -38,7 +38,7 @@
      *
      * @param numValues Number of values of the function to compute.
      * @param cache Cached values.
-     * @throw IllegalArgumentException if {@code n < 0}.
+     * @throw CombinatoricsException if {@code n < 0}.
```
aherbert, line 41: `CombinatoricsException` is package private; it should not be exposed in the javadoc. It extends IllegalArgumentException so the javadoc is still valid.

commons-numbers-primes/src/main/java/org/apache/commons/numbers/primes/SmallPrimes.java
```diff
@@ -108,11 +108,11 @@ equivalence classes will have to hold (2-1)*(3-1)*(5-1)*(7-1)*(11-1) = 480
         of all integers.
          */
         final Set<Integer> primeNumbers = new HashSet<>();
-        primeNumbers.add(Integer.valueOf(2));
-        primeNumbers.add(Integer.valueOf(3));
-        primeNumbers.add(Integer.valueOf(5));
-        primeNumbers.add(Integer.valueOf(7));
-        primeNumbers.add(Integer.valueOf(11));
+        primeNumbers.add(2);
```
aherbert, line 111: The current code makes the boxing explicit. Please avoid auto-boxing.

arturobernalg: Changed

commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/FactorialDouble.java
arturobernalg: Changed

commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/LogFactorial.java
```diff
@@ -41,7 +41,7 @@
      *
      * @param numValues Number of values of the function to compute.
      * @param cache Cached values.
-     * @throw IllegalArgumentException if {@code n < 0}.
+     * @throw CombinatoricsException if {@code n < 0}.
```
aherbert, line 44: You have left this `CombinatoricsException` exposed in the API.

arturobernalg: Changed
TY

aherbert:
This is now only simplifying boolean expressions. Can you update the commit message to reflect that and then correct the formatting as per #103

arturobernalg:
> This is now only simplifying boolean expressions. Can you update the commit message to reflect that and then correct the formatting as per #103

HI @aherbert 
Done
TY

aherbert:
You changed the commit message but did not fix formatting to change 8 spaces to 4.

PR #105, opened 2021-10-08
NUMBERS-167: Allow precomputation of regularized gamma arguments

aherbert:


aherbert:
Out of date and requires large refactoring around the new Gamma function implementation.

PR #114, opened 2022-06-26
Some minors improvements.

arturobernalg:
* Use Math.min
* Use System.arraycopy

### DIFF
--- a/commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/LogFactorial.java
+++ b/commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/LogFactorial.java
@@ -55,15 +55,14 @@ public final class LogFactorial {
         int endCopy;
         if (cache == null || cache.length <= beginCopy) {
             endCopy = beginCopy;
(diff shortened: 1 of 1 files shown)


commons-numbers-combinatorics/src/main/java/org/apache/commons/numbers/combinatorics/LogFactorial.java
```diff
@@ -55,15 +55,14 @@ private LogFactorial(int numValues,
...
-            endCopy = numValues;
+            endCopy = Math.min(cache.length, numValues);
         }
 
+
         // Copy available values.
-        for (int i = beginCopy; i < endCopy; i++) {
-            logFactorials[i] = cache[i];
+        if (endCopy - beginCopy > 0) {
+            System.arraycopy(cache, beginCopy, logFactorials, beginCopy, endCopy - 2);
```
aherbert, line 65: Should this be: length = endCopy - beginCopy, i.e.
```Java
System.arraycopy(cache, beginCopy, logFactorials, beginCopy, endCopy - beginCopy);
```

aherbert (changes requested):

PR #117, opened 2022-07-15
Numbers 188.refactor complex log function

sumanth-rajkumar:
This PR refactors Complex class using functional interfaces and static methods. 
This allows reuse of the refactored static functions for Operations on List/Matrices of Complex Numbers (NUMBERS-186)

The changes retain binary backward compatibility.

Summary of changes

1) Introduced following interfaces  for Complex Number Operations

 ComplexUnaryOperator - unary operations on complex number represented as double real and imaginary parts
 ComplexConstructor - interface to create a generic complex result from double real and imaginary parts

2) As a first step, refactored log and log10 instance methods of Complex class as static functions in ComplexFunctions
The static functions use the functional interface signatures described above


commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConstructor.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.numbers.complex;
+
+/**
+ * Define a constructor for a Complex.
```
aherbert, line 21: This functional interface requires more details. It should highlight which is the function method. See my example for the ComplexUnaryOperator.

All new public files should have `@since 1.1` tags.

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConstructor.java
```diff
@@ -0,0 +1,36 @@
...
+public interface ComplexConstructor<R> {
+
+    /**
+     * Represents a function that accepts real and imaginary part of complex number and returns an object.
+     * @param real real part
+     * @param imaginary imaginary part
+     * @return R
+     */
+    R apply(double real, double imaginary);
+
```
aherbert, line 34: Remove empty line

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConstructor.java
```diff
@@ -0,0 +1,36 @@
...
+ * @param <R> Generic
+ */
+@FunctionalInterface
+public interface ComplexConstructor<R> {
+
+    /**
+     * Represents a function that accepts real and imaginary part of complex number and returns an object.
+     * @param real real part
+     * @param imaginary imaginary part
+     * @return R
```
aherbert, line 31: the object encapsulating the complex result

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConstructor.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.apache.commons.numbers.complex;
+
+/**
+ * Define a constructor for a Complex.
+ * @param <R> Generic
+ */
+@FunctionalInterface
+public interface ComplexConstructor<R> {
```
darkma773r, line 25: The word "constructor" has a very specific meaning in Java. Perhaps we could use "ComplexFactory" here instead?

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
```diff
@@ -0,0 +1,813 @@
...
+     * <a href="http://www.iso-9899.info/wiki/The_Standard">ISO C Standard</a>, Annex G,
+     * in calculating the returned value using the {@code atan2(y, x)} method for complex
+     * \( x + iy \).
+     *
+     * @param r real part of complex number
+     * @param i imaginary part of complex number
+     * @return The argument of this complex number.
+     * @see Math#atan2(double, double)
+     */
+    public static double arg(double r, double i) {
```
aherbert, line 127: This should be called by Complex for consistency.

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
```diff
@@ -0,0 +1,813 @@
...
+     * @see Double#isInfinite(double)
+     */
+    private static boolean isInfinite(double real, double imaginary) {
+        return Double.isInfinite(real) || Double.isInfinite(imaginary);
+    }
+
+    /**
+     * Returns the
+     * <a href="http://mathworld.wolfram.com/NaturalLogarithm.html">
+     * natural logarithm</a> of this complex number using its real and imaginary parts.
```
aherbert, line 149: References to `this complex number` should be updated to `the complex number`

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
```diff
@@ -0,0 +1,813 @@
...
+     * @param logOfeOver2 The log function applied to e, then divided by 2.
+     * @param logOf2 The log function applied to 2.
+     * @param real real part of complex number
+     * @param imaginary imaginary part of complex number
+     * @param constructor Constructor for the returned complex.
+     * @param <R> generic
+     * @return The logarithm of this complex number.
+     */
+    private static <R> R log(DoubleUnaryOperator log, double logOfeOver2, double logOf2,
+                                     double real, double imaginary, ComplexConstructor<R> constructor) {
```
aherbert, line 246: Fix the trailing argument indentation

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/CStandardTest.java
```diff
@@ -508,6 +593,146 @@ private static void assertComplex(Complex z,
...
+
+        // Test the operation
+        final Complex c = operation.apply(z);
+        final ComplexNumber c1 = operation2.apply(z.getReal(), z.getImaginary(), ComplexNumber::new);
+        final Complex t1 = sign.removeSign(c);
+        final Complex t2 = sign.removeSign(expected);
+        if (!equals(t1.getReal(), t2.getReal()) ||
+            !equals(t1.getImaginary(), t2.getImaginary())) {
+            Assertions.fail(
+                String.format("Operation failed (z=%s). Expected: %s but was: %s (Unspecified sign = %s)",
```
aherbert, line 712: This formatting message includes c1 as an argument but the string format does not require it.

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexNumber.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Cartesian representation of a complex number. The complex number is expressed
+ * in the form \( a + ib \) where \( a \) and \( b \) are real numbers and \( i \)
+ * is the imaginary unit which satisfies the equation \( i^2 = -1 \). For the
+ * complex number \( a + ib \), \( a \) is called the <em>real part</em> and
+ * \( b \) is called the <em>imaginary part</em>.
+ */
+class ComplexNumber {
+
+    private final double real;
```
aherbert, line 29: Comment these for completeness.

aherbert (changes requested):
Hi @sumanth-rajkumar 

Thanks for the new PR. There are many items to address. Please also see my comments from the previous PR as some of them are the same (you have copied code from that PR without addressing the previous feedback).

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConstructor.java
aherbert: I agree the term constructor is not ideal. The function is acting as both a Consumer (to receive the complex result) and optionally a Function (to convert (a +ib) to z). In both cases the operation is a terminating operation where the primitive (r, i) pair finally gets used and potentially converted to an object. So there is a lot to cover in a name. Switching to Factory implies that the function will always make something (as does constructor). However some use cases could return `Void`. So I think this also needs to be captured in the name. Perhaps ComplexResult?

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/Complex.java
```diff
@@ -2905,224 +2776,7 @@ private static Complex atanh(final double real, final double imaginary,
      * @return {@code x^2 + y^2 - 1}.
      */
     private static double x2y2m1(double x, double y) {
```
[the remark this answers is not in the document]
sumanth-rajkumar: This method is also used in log which is why I moved it to ComplexFunctions. Should I just make it public in Complex and use it from there in ComplexFunctions or is there something else I can do?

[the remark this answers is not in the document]
aherbert: I did not see that. Just make it package private and add a TODO note to make it private in the future.

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConstructor.java
darkma773r: What do you think of using the name pair "source"/"sink". I could then picture the following:
```java
// interface for obtaining complex values
public interface ComplexSource {
    double getReal();
    double getImaginary();

    default real() { return getReal(); }
    default imag() { return getImaginary(); }
}

// interface for accepting complex values, optionally returning a result
@FunctionalInterface
public interface ComplexSink<R> {
    R accept(double real, double imaginary);
}
```
These interfaces could then be used as follows:
```java
// existing Complex class
public final class Complex implements ComplexSource {
    // ...
}

// TBD complex value container class
public class ComplexBuffer implements ComplexSource, ComplexSink<ComplexBuffer> {
    // indexing methods similar to java.nio.Buffer
    public int position();
    public ComplexBuffer position(int pos);

    // get the values at the current position
    public double getReal() { ... }
    public double getImaginary() { ... }

    // set the values at the current position
    ComplexBuffer accept(double real, double imaginary) {
        // ...
        return this;
    }
}
```

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexTest.java
```diff
@@ -1419,6 +1419,9 @@ void testLog10() {
             final Complex z = Complex.ofCartesian(rng.nextDouble() * 2, rng.nextDouble() * 2);
             final Complex lnz = z.log();
             final Complex log10z = z.log10();
+            final ComplexNumber log10z2 = ComplexFunctions.log10(z.getReal(), z.getImaginary(), ComplexNumber::new);
```
aherbert, line 1422: There is no test for ComplexFunctions.log.

In this case it may be convenient to add a method to TestUtils to assert two operations are the same, then return the result. You can then do further assertions on the result.

```Java
    static Complex assertSame(Complex c,
                              UnaryOperator<Complex> operation1,
                              ComplexUnaryOperator<ComplexNumber> operation2,
                              String name) {
        final Complex z = operation1.apply(c);
        // Test operation2 produces the exact same result
        operation2.apply(c.real(), c.imag(), (x, y) -> {
            Assertions.assertEquals(z.real(), x, () -> name + " real");
            Assertions.assertEquals(z.imag(), y, () -> name + " imaginary");
            return null;
        });
        return z;
    }

Complex lnz = TestUtils.assertSame(z, Complex::log, ComplexFunctions::log, "log");
Complex ln10z = TestUtils.assertSame(z, Complex::log10, ComplexFunctions::log10, "log10");
// Test lnz ...
```

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
```diff
@@ -0,0 +1,584 @@
...
+import java.util.function.DoubleUnaryOperator;
+
+/**
+ * Cartesian representation of a complex number. The complex number is expressed
+ * in the form \( a + ib \) where \( a \) and \( b \) are real numbers and \( i \)
+ * is the imaginary unit which satisfies the equation \( i^2 = -1 \). For the
+ * complex number \( a + ib \), \( a \) is called the <em>real part</em> and
+ * \( b \) is called the <em>imaginary part</em>.
+ *
+ * <p>This class is immutable, it is to change from an OO representation of a Complex number
```
aherbert, line 28: The class is a utility class with only static methods so immutability is not relevant. I think just remove this paragraph.

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
```diff
@@ -0,0 +1,584 @@
...
+     * <p>The implementation is based on the method described in:</p>
+     * <blockquote>
+     * T E Hull, Thomas F Fairgrieve and Ping Tak Peter Tang (1994)
+     * Implementing complex elementary functions using exception handling.
+     * ACM Transactions on Mathematical Software, Vol 20, No 2, pp 215-244.
+     * </blockquote>
+     *
+     * @param real Real part \( a \) of the complex number \(a +ib \).
+     * @param imaginary Imaginary part \( b \) of the complex number \(a +ib \).
+     * @param constructor Terminating consumer for the complex result, used to construct a result of type {@code R}.
```
aherbert, line 209: Rename `constructor` to `action`. Move the javadoc description from the return tag to this tag. The return tag can be made more generic as you do not know what the object will represent.
```
@param action Consumer for the natural logarithm of the complex number

@return the object returned by the supplied action
...


```

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/Complex.java
```diff
@@ -2904,7 +2782,8 @@ private static Complex atanh(final double real, final double imaginary,
      * @param y the y value
      * @return {@code x^2 + y^2 - 1}.
      */
-    private static double x2y2m1(double x, double y) {
+    //TODO - make it private in future
```
aherbert, line 2785: Since the intent is to eventually move all these private functions to ComplexFunctions, then please start the process now. Make any function that is currently also required in Complex package-private and add the TODO note to make private in CompexFunctions in the future.

It may be convenient to leave the functions locally as pass through functions:
```java
private static double x2y2m1(double x, double y) {
    return ComplexFunctions.x2y2m1(x, y);
}
```
This will prevent having to update all the call sites in Complex and minimise the diff.

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
```diff
@@ -0,0 +1,584 @@
...
+        //
+        // 3. An alteration is done here to add an 'else if' instead of a second
+        // 'if' statement. Thus you cannot scale down and up at the same time.
+        //
+        // 4. There is no use of the absolute double value. The magnitude comparison is
+        // performed using the long bit representation. The computation x^2+y^2 is
+        // insensitive to the sign bit. Thus use of Math.abs(double) is only in edge-case
+        // branches.
+        //
+        // 5. The exponent different to ignore the smaller component has changed from 60 to 54.
```
aherbert, line 444: Correct original typo to `difference`

aherbert (changes requested):
This is much improved.

Can you rename the ComplexResult to ComplexSink and remove use of the word constructor and constructs from all of ComplexFunctions. The result argument should be named `action`.

I think all the assertion functions that accept two operations should in javadoc:

- describe the assertion to the expected
- state the two operations are exactly the same

I've suggested using an assertSame function in TestUtils to test the two functions are the same and return the Complex. This should simplify converting all of ComplexTest to test both implementations when the assertions on the result `z` are custom.

It may be useful for the other assertion functions in CReferenceTest etc do it this way too. This essentially offloads the check the two functions are exactly equal elsewhere. The assertions can then just test the result complex `z` satisfies the equals requirements (essentially just reusing the current test code).

commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexFunctions.java
sumanth-rajkumar: Which paragraph do you mean? Do you mean to just leave the part of changing from OO representation and take out everything else in that paragraph?

aherbert: Remove the paragraph starting within `<p>This class is immutable, ...`. It is about 3 lines of the javadoc.

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/CStandardTest.java
```diff
@@ -300,6 +300,103 @@ private static void assertConjugateEquality(Complex z,
         }
     }
 
+    /**
+     * Assert the operation on the complex number satisfies the conjugate equality.
+     * Assert the operation on the complex number is exactly equal to the operation on
```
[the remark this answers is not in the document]
aherbert: Move the new javadoc you added. The method should essentially be documented as before. What you have done is add a second operation as an argument. This should be documented, after the existing javadoc, as tested to exactly match the result of the first operation. I said put it before the <h2> section as that is a natural break in the javadoc. Below that is an explanation of conjugates.

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/CStandardTest.java
```diff
@@ -300,6 +300,107 @@ private static void assertConjugateEquality(Complex z,
...
+     * @param operation2 the operation on the complex real and imaginary parts
+     * @param sign the sign specification
+     */
+    private static void assertConjugateEquality(Complex z,
+        UnaryOperator<Complex> operation1,
+        ComplexUnaryOperator<ComplexNumber> operation2,
+        UnspecifiedSign sign) {
+
+        final Complex zConj = z.conj();
+        final Complex c1 = operation1.apply(zConj);
```
aherbert, line 383: Replace with:
```Java
final Complex c1 = TestUtils.assertSame(zConj, operation1, operation2, "");
final Complex c2 = TestUtils.assertSame(z, operation1, operation2, "").conj();
```

Then you do not need to generate ComplexNumber cn1 and cn2.

This possibly requires a name argument. It is missing from the original tests. I cannot recall why and so can you look into adding it so we know the conjugate equality that failed.

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/CStandardTest.java
```diff
@@ -508,6 +609,109 @@ private static void assertComplex(Complex z,
...
+     * Assert the operation on the complex number is equal to the expected value.
+     * If the imaginary part is not NaN the operation must also satisfy the conjugate equality.
+     *
+     * <pre>
+     * op(conj(z)) = conj(op(z))
+     * </pre>
+     *
+     * <p>The results must be binary equal. This includes the sign of zero.
+     *
+     * Assert the operation on the complex number is <em>exactly</em> equal to the operation on
```
aherbert, line 622: New paragraph requires `<p>` tag

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/CStandardTest.java
```diff
@@ -300,6 +300,107 @@ private static void assertConjugateEquality(Complex z,
...
+     * Assert the operation on the complex number satisfies the conjugate equality.
+     *
+     * <pre>
+     * op(conj(z)) = conj(op(z))
+     * </pre>
+     *
+     * <p>The results must be binary equal; the sign of the complex number is first processed
+     * using the provided sign specification.
+     *
+     * Assert the operation on the complex number is <em>exactly</em> equal to the operation on
```
aherbert, line 369: Missing `<p>`

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexEdgeCaseTest.java
```diff
@@ -81,6 +81,57 @@ private static void assertComplex(double a, double b,
...
+    }
+
+    /**
+     * Assert the operation on the complex number is equal to the expected value.
+     *
+     * <p>The results are considered equal within the provided units of least
+     * precision. The maximum count of numbers allowed between the two values is
+     * {@code maxUlps - 1}.
+     *
+     * Assert the operation on the complex number is <em>exactly</em> equal to the operation on
```
aherbert, line 114: Missing `<p>`

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexEdgeCaseTest.java
```diff
@@ -81,6 +81,57 @@ private static void assertComplex(double a, double b,
...
+     * complex real and imaginary parts.
+     *
+     * @param a Real part.
+     * @param b Imaginary part.
+     * @param name The operation name.
+     * @param operation1 the operation on the Complex object.
```
aherbert, line 120: Drop `the`, capitalize, etc

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexNumber.java
```diff
@@ -0,0 +1,76 @@
...
+    /** The imaginary part. */
+    private final double imaginary;
+
+    /**
+     * Constructor representing a complex number by its real and imaginary parts.
+     * Takes in real and imaginary and sets it to this complex number's real and imaginary
+     *
+     * @param real Real part \( a \) of the complex number \( (a +ib \).
+     * @param imaginary Imaginary part \( b \) of the complex number \( (a +ib \).
+     *
```
aherbert, line 40: Extra line

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexNumber.java
```diff
@@ -0,0 +1,76 @@
...
+class ComplexNumber {
+
+    /** The real part. */
+    private final double real;
+    /** The imaginary part. */
+    private final double imaginary;
+
+    /**
+     * Constructor representing a complex number by its real and imaginary parts.
+     * Takes in real and imaginary and sets it to this complex number's real and imaginary
```
aherbert, line 36: Second sentence is not required. I think it may be left over from when this was a method not a constructor.

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/ComplexNumber.java
```diff
@@ -0,0 +1,76 @@
...
+        this.imaginary = imaginary;
+    }
+
+    /**
+     * Creates a conjugated complex number given the real and imaginary parts,
+     * that is for the argument (a + ib), returns (a - ib).
+     *
+     * @param real Real part \( a \) of the complex number \( (a +ib \).
+     * @param imaginary Imaginary part \( b \) of the complex number \( (a +ib \).
+     * @return {@code ComplexNumber} object.
```
aherbert, line 53: `@return conjugated complex number`

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/TestUtils.java
```diff
@@ -67,6 +68,19 @@ public static void assertSame(Complex expected, Complex actual) {
...
 
+    /**
+     * Verifies that real and imaginary parts of the Complex and ComplexNumber arguments are
+     * exactly the same as defined by {@link Double#compare(double, double)}. Also
+     * ensures that NaN / infinite components match.
+     *
+     * @param expected the expected value
+     * @param actual the actual value
+     */
+    public static void assertSame(Complex expected, ComplexNumber actual) {
```
aherbert, line 79: Do we need this method? If so then add a message to the assertions for "real" and "imaginary"

commons-numbers-complex/src/test/java/org/apache/commons/numbers/complex/TestUtils.java
```diff
@@ -387,4 +401,28 @@ private static String preprocessTestData(String line, TestDataFlagOption option,
...
+     * complex real and imaginary parts.
+     *
+     * @param c The input complex number.
+     * @param operation1 the operation on the Complex object
+     * @param operation2 the operation on the complex real and imaginary parts
+     * @param name the operation name
+     * @return the resulting complex number from the given operation
+     */
+    static Complex assertSame(Complex c,
+                             UnaryOperator<Complex> operation1,
```
aherbert, line 416: Indent 1 more space

aherbert (changes requested):
Hi @sumanth-rajkumar 

I have suggested a few changes. Please contact me if you need more details.

PR #122, opened 2022-09-03
NUMBERS-186: Operations on Lists of Complex Numbers

sumanth-rajkumar:
This PR introduces support for complex numbers in the complex package to allow operations to be performed on lists of complex numbers. The list can be efficiently iterated to apply all the operations in-place supported by the Complex class.

Summary of changes

ComplexListFunctionsTest - unit tests for complex operations on ComplexList


commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -19,16 +19,15 @@
 
 import org.apache.commons.numbers.complex.Complex;
 import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.junit.jupiter.api.Test;
```
aherbert, line 24: Reorganising imports should not be done

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListFunctionsTest.java
```diff
@@ -0,0 +1,69 @@
...
+        List<Complex> objectList = generateList(10);
+        double factor = 2;
+        ComplexList actualList = new ComplexList();
+        actualList.addAll(objectList);
+        objectList.replaceAll(c -> c.pow(factor));
+        actualList.replaceAll((x, y, action) -> ComplexFunctions.pow(x, y, factor, action));
+        Assertions.assertEquals(objectList, actualList);
+    }
+
+    private List<Complex> generateList(int size) {
```
aherbert, line 65: Can be static. Should have a comment indicating what the function generates

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -322,4 +325,79 @@ private String outOfBoundsMsg(int index) {
...
+        }
+    }
+
+    /**
+     * Replaces each element of the list with the result of applying the operator to that element.
+     * @param operator The operator to apply to each element.
+     */
+    public void replaceAll(ComplexUnaryOperator<Void> operator) {
+        ComplexCursor cursor = new ComplexCursor();
+        while (cursor.next()) {
```
aherbert, line 397: This still advances the index outside of the cursor. I think the cursor should have that responsibility.

The cursor could be changed to use the action in a tryAdvance(action) pattern. Alternatively if the cursor will not be reused then it could be removed and the logic placed directly here. It would eliminate the object allocation and simplify the loop logic.

```Java
final double[] parts = ...;
final int m = size;
for (int i = 0; i < m; i++) {
    final int index = i << 1;
    operator.apply(parts[index], parts[index + 1], (x, y) -> { parts[index] = x; parts[index + 1] = y; });
}
// check for comodification
```
Doing it this way uses a local copy of the list and size for efficiency.

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListFunctionsTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.commons.numbers.complex.Complex;
+import org.apache.commons.numbers.complex.ComplexFunctions;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
+
+public class ComplexListFunctionsTest {
```
aherbert, line 29: These tests can be moved to ComplexListTest, or at least the test class should be commented to state what tests are in this that are not in ComplexListTest.

aherbert (changes requested):

aherbert:
Closed by 603a17f54e730e518d0bd0f2d9cef553a5e72808

PR #123, opened 2022-09-06
NUMBERS-186 added additional complex list operations

sumanth-rajkumar:
This PR introduces additional operations that can be done on ComplexList. 

Summary of changes

ComplexListTest - unit tests for complex additional operations on ComplexList
ComplexConsumer -  interface that acts as a consumer for the complex element's double real and imaginary parts

### DIFF
--- a/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
+++ b/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
@@ -210,8 +210,7 @@ public class ComplexList extends AbstractList<Complex> {
      */
     public void setReal(int index, double real) {
         rangeCheck(index);
-        final int i = index << 1;
-        realAndImagParts[i] = real;
+        realAndImagParts[index << 1] = real;
     }
 
     /**
@@ -223,8 +222,7 @@ public class ComplexList extends AbstractList<Complex> {
      */
     public void setImaginary(int index, double imaginary) {
         rangeCheck(index);
-        final int i = index << 1;
-        realAndImagParts[i + 1] = imaginary;
+        realAndImagParts[(index << 1) + 1] = imaginary;
     }
 
     /**
--- a/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
+++ b/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
@@ -29,17 +29,17 @@ import java.util.Objects;
 
 /**
  * Resizable-double array implementation of the List interface. Implements all optional list operations,
- * and permits all elements. In addition to implementing the List interface,
+ * and permits all complex numbers. In addition to implementing the List interface,
  * this class provides methods to manipulate the size of the array that is used internally to store the list.
  *
- * <p>Each ComplexList instance has a capacity. The capacity is half the size of the double array used to store the elements
- * in the list. As elements are added to an ComplexList, its capacity grows automatically.
- * The complex number is stored using an interleaved format and so the maximum number of elements that may be added is
+ * <p>Each ComplexList instance has a capacity. The capacity is half the size of the double array used to store the complex numbers
+ * in the list. As complex numbers are added to an ComplexList, its capacity grows automatically.
+ * The complex number is stored using an interleaved format and so the maximum number of complex numbers that may be added is
  * approximately 2^30. This is half the maximum capacity of java.util.ArrayList.
  * The memory usage is more efficient than using a List of Complex objects as the underlying numbers are not stored
  * using instances of Complex.</p>
  *
- * <p>An application can increase the capacity of an ComplexList instance before adding a large number of elements
+ * <p>An application can increase the capacity of an ComplexList instance before adding a large number of complex numbers
  * using the ensureCapacity operation. This may reduce the amount of incremental reallocation.</p>
  */
 public class ComplexList extends AbstractList<Complex> {
@@ -56,7 +56,7 @@ public class ComplexList extends AbstractList<Complex> {
     private static final int MAX_CAPACITY = MAX_ARRAY_SIZE / 2;
 
     /**
-     * error in case of allocation above max capacity.
+     * Error in case of allocation above max capacity.
      */
     private static final String OOM_ERROR_STRING = "cannot allocate capacity %s greater than max " + MAX_CAPACITY;
 
@@ -75,7 +75,7 @@ public class ComplexList extends AbstractList<Complex> {
     private static final String INDEX_MSG = "Index: ";
 
     /**
-     * The double array buffer into which the elements of the ComplexList are stored.
+     * The double array buffer into which the complex numbers of the ComplexList are stored.
      */
     private double[] realAndImagParts;
 
@@ -116,7 +116,7 @@ public class ComplexList extends AbstractList<Complex> {
     /**
      * Checks if the given index is in range.
      *
-     * @param index Index of the element to range check.
+     * @param index Index of the complex number to range check.
      * @throws IndexOutOfBoundsException if index isn't within the range.
      */
     private void rangeCheck(int index) {
@@ -128,7 +128,7 @@ public class ComplexList extends AbstractList<Complex> {
     /**
      * A version of rangeCheck used by add and addAll.
      *
-     * @param index Index of the element to range check.
+     * @param index Index of the complex number to range check.
      * @throws IndexOutOfBoundsException if index isn't within the range of list.
      */
     private void rangeCheckForInsert(int index) {
@@ -154,32 +154,30 @@ public class ComplexList extends AbstractList<Complex> {
     /**
      * Gets the real part \( a \) of the complex number \( (a + i b) \) at the indexed position of the list.
      *
-     * @param index Index of the element's real part to get.
+     * @param index Index of the complex number.
      * @return The real part.
      */
     public double getReal(int index) {
         rangeCheck(index);
-        final int i = index << 1;
-        return realAndImagParts[i];
+        return realAndImagParts[index << 1];
     }
 
     /**
-     * Gets the imaginary part \( b \) of this complex number \( (a + i b) \).
+     * Gets the imaginary part \( b \) of the complex number \( (a + i b) \) at the indexed position of the list.
      *
-     * @param index Index of the element's imaginary part to get.
+     * @param index Index of the complex number.
      * @return The imaginary part.
      */
     public double getImaginary(int index) {
         rangeCheck(index);
-        final int i = index << 1;
-        return realAndImagParts[i + 1];
+        return realAndImagParts[(index << 1) + 1];
     }
 
     /**
-     * Replaces the element at the specified position in this list with the specified element's
+     * Replaces the complex number at the specified position in the list with the specified complex number's
      * real and imaginary parts. No range checks are done.
      *
-     * @param index Index of the element to replace.
+     * @param index Index of the complex number to replace.
      * @param real Real part \( a \) of the complex number \( (a +ib) \).
      * @param imaginary Imaginary part \( b \) of the complex number \( (a +ib) \).
      */
@@ -192,22 +190,22 @@ public class ComplexList extends AbstractList<Complex> {
     /**
      * {@inheritDoc}
      *
-     * @param element Complex element to be set.
+     * @param number Complex number to be set.
      */
     @Override
-    public Complex set(int index, Complex element) {
+    public Complex set(int index, Complex number) {
         rangeCheck(index);
         final int i = index << 1;
         final Complex oldValue = Complex.ofCartesian(realAndImagParts[i], realAndImagParts[i + 1]);
-        setNoRangeCheck(index, element.getReal(), element.getImaginary());
+        setNoRangeCheck(index, number.getReal(), number.getImaginary());
         return oldValue;
     }
 
     /**
-     * Replaces the complex element's real part at the specified position
-     * in this list with the specified real element.
+     * Replaces the complex number's real part at the specified position
+     * in the list with the specified real number.
      *
-     * @param index Index of the element's real part to replace.
+     * @param index Index of the complex number.
      * @param real Real part \( a \) of the complex number \( (a +ib) \).
      */
     public void setReal(int index, double real) {
@@ -217,9 +215,10 @@ public class ComplexList extends AbstractList<Complex> {
     }
 
     /**
-     * Replaces the complex element's imaginary part at the specified position
-     * in this list with the specified imaginary element.
-     * @param index Index of the element's imaginary part to replace.
+     * Replaces the complex number's imaginary part at the specified position
+     * in the list with the specified imaginary number.
+     *
+     * @param index Index of the complex number.
      * @param imaginary Imaginary part \( b \) of the complex number \( (a +ib) \).
      */
     public void setImaginary(int index, double imaginary) {
@@ -229,8 +228,8 @@ public class ComplexList extends AbstractList<Complex> {
     }
 
     /**
-     * Returns an array containing all the complex element's real parts in
-     * the list in proper sequence (from first to last element).
+     * Returns an array containing all the complex number's real parts in
+     * the list in proper sequence (from first to last number).
      *
      * @return Array of real parts.
      */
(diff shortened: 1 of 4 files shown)


commons-numbers-complex/src/main/java/org/apache/commons/numbers/complex/ComplexConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@FunctionalInterface
+public interface ComplexConsumer {
+
+    /**
+     * Represents a function that accepts real and imaginary part of complex number and returns no result.
+     * @param real Real part \( a \) of the complex number \( (a + ib) \).
+     * @param imaginary Imaginary part \( b \) of the complex number \( (a + ib) \).
+     */
+    void apply(double real, double imaginary);
```
aherbert, line 36: java.util.function.Consumer has the method `accept`

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -149,6 +151,30 @@ public Complex get(int index) {
...
 
+    /**
+     * Gets the real part \( a \) of the complex number \( (a + i b) \) at the indexed position of the list.
+     *
+     * @param index Index of the element's real part to get.
+     * @return The real part.
+     */
+    public double getReal(int index) {
+        rangeCheck(index);
+        final int i = index << 1;
```
aherbert, line 162: No need for this variable, just perform the index doubling inline

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -149,6 +151,30 @@ public Complex get(int index) {
         return Complex.ofCartesian(realAndImagParts[i], realAndImagParts[i + 1]);
     }
 
+    /**
+     * Gets the real part \( a \) of the complex number \( (a + i b) \) at the indexed position of the list.
+     *
+     * @param index Index of the element's real part to get.
```
aherbert, line 157: `Index of the complex number.`

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -177,6 +203,63 @@ public Complex set(int index, Complex element) {
         return oldValue;
     }
 
+    /**
+     * Replaces the complex element's real part at the specified position
```
aherbert, line 207: change `element` to `number`. Element is for a generic list; since this is typed to Complex then all element occurrences should be changed to `complex number`

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -177,6 +203,63 @@ public Complex set(int index, Complex element) {
...
+    public void setReal(int index, double real) {
+        rangeCheck(index);
+        final int i = index << 1;
+        realAndImagParts[i] = real;
+    }
+
+    /**
+     * Replaces the complex element's imaginary part at the specified position
+     * in this list with the specified imaginary element.
+     * @param index Index of the element's imaginary part to replace.
```
aherbert, line 222: Add an empty line

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -177,6 +203,63 @@ public Complex set(int index, Complex element) {
...
+     * Returns an array containing all the complex element's real parts in
+     * the list in proper sequence (from first to last element).
+     *
+     * @return Array of real parts.
+     */
+    double[] toArrayReal() {
+        final int length = size;
+        double[] real = new double[length];
+        for (int i = 0; i < length; i++) {
+            final int index = i << 1;
```
aherbert, line 241: Again this `index` can be inlined

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -348,4 +433,25 @@ public void replaceAll(ComplexUnaryOperator<Void> operator) {
...
+    /**
+     * Replaces each element of the list with the result of applying the action to that element.
+     *
+     * @param action The action to apply to each element.
+     */
+    public void forEach(ComplexConsumer action) {
+        Objects.requireNonNull(action);
+        final double[] parts = this.realAndImagParts;
+        final int m = size;
+        final int expectedModCount = modCount;
```
aherbert, line 446: The mod count is increased when the list is structurally modified. Here no modifications are occurring so this can be removed.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -348,4 +433,25 @@ public void replaceAll(ComplexUnaryOperator<Void> operator) {
         }
         modCount++;
     }
+
+    /**
+     * Replaces each element of the list with the result of applying the action to that element.
```
aherbert, line 438: This javadoc is wrong. Look at the javadoc for `forEach(Consumer<Complex>)` and adapt as appropriate.

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
         Assertions.assertEquals(objectList, actualList);
     }
 
+    @Test
+    void testReplaceAllComplexConsumer() {
```
aherbert, line 228: `testForEach...`

This test can be parameterized for size. This will eliminate duplication:
```Java
@ParameterizedTest
@ValuesSource(ints = {0, 10})
void testForEachComplexConsumer(int size) {
    List<Complex> data1 = generateList(size);
```

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
...
+        expected2.addAll(data2);
+        expected2.forEach((real, imaginary) -> actual2.add(Complex.ofCartesian(real, imaginary)));
+        Assertions.assertEquals(expected2, actual2);
+    }
+
+    @Test
+    void testGetRealAndImaginary() {
+        ComplexList list = new ComplexList();
+        list.add(Complex.ofCartesian(42, 13));
+        list.addAll(1, list);
```
aherbert, line 250: Please stop creating larger lists using duplication. The list elements should be unique. Create the list using random complex numbers or a sequence.

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
...
+        list.addAll(list);
+        list.set(2, Complex.ofCartesian(200, 1));
+        for (int i = 0; i < list.size(); i++) {
+            Assertions.assertEquals(list.get(i).getReal(), list.getReal(i));
+        }
+        for (int i = 0; i < list.size(); i++) {
+            Assertions.assertEquals(list.get(i).getImaginary(), list.getImaginary(i));
```
aherbert, line 257: This can be placed in the same loop as above. Add a simple message to the assertion, e.g. "real".

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
...
+
+    @Test
+    void testSetRealAndImaginary() {
+        ComplexList list = new ComplexList();
+        list.add(Complex.ofCartesian(42, 13));
+        list.addAll(1, list);
+        list.addAll(list);
+        list.setReal(2, 200);
+        list.setImaginary(2, 1);
+        Assertions.assertEquals(Complex.ofCartesian(200, 1), list.get(2));
```
aherbert, line 269: Test set at all positions:
```Java
for (int i = 0; i < list.size(); i++) {
    final double value = Math.PI * i;
    list.setReal(i, value);
    Assertions.assertEquals(value, list.get(i).getReal());
}
```
Also test set/get with invalid index positions (-1, size, size+1).

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
...
+    }
+
+    @Test
+    void testToArrayRealAndImaginary() {
+        ComplexList list = new ComplexList();
+        list.add(Complex.ofCartesian(42, 13));
```
aherbert, line 275: Create a better random list.

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
...
+        list.addAll(1, list);
+        list.addAll(list);
+        list.set(2, Complex.ofCartesian(200, 1));
+
+        double[] expectedReal = {42, 42, 200, 42};
```
aherbert, line 280: double[] expected = list.stream().mapToDouble(Complex::getReal).toArray();

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,6 +224,71 @@ void testReplaceAllComplexScalarFunction() {
...
+        double[] actualReal = list.toArrayReal();
+        for (int i = 0; i < expectedReal.length; i++) {
```
aherbert, line 282: Use `Assertions.assertArrayEquals`

aherbert (changes requested):
The forEach has been copied from replaceAll without removing the concurrent modification check or updating the javadoc. The test is also incorrectly named.

For your javadoc can you make sure you do not use the term element; here we should use complex number.

The tests should use data where each item in the list is unique. Index out of bound exceptions should be tested, especially at the upper size limit as without a range check the get methods can return stale/invalid data and the set methods may silently not error and write to an unused part of the backing array.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -166,20 +190,76 @@ private void setNoRangeCheck(int index, double real, double imaginary) {
...
+    /**
+     * Replaces the complex number's imaginary part at the specified position
+     * in the list with the specified imaginary number.
+     *
+     * @param index Index of the complex number.
+     * @param imaginary Imaginary part \( b \) of the complex number \( (a +ib) \).
+     */
+    public void setImaginary(int index, double imaginary) {
+        rangeCheck(index);
+        final int i = index << 1;
```
aherbert, line 226: This index i should be inlined.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -166,20 +190,76 @@ private void setNoRangeCheck(int index, double real, double imaginary) {
...
+    /**
+     * Replaces the complex number's real part at the specified position
+     * in the list with the specified real number.
+     *
+     * @param index Index of the complex number.
+     * @param real Real part \( a \) of the complex number \( (a +ib) \).
+     */
+    public void setReal(int index, double real) {
+        rangeCheck(index);
+        final int i = index << 1;
```
aherbert, line 213: This index i should be inlined.

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -152,9 +153,9 @@ void testAddIndexOutOfBoundExceptions() {
 
     @Test
     void testRemoveIndexOutOfBoundExceptions() {
+        List<Complex> objectList = generateList(2);
```
aherbert, line 156: Since a lot of the tests now use generateList() then addAll to a new ComplexList instance this should be refactored into a method: `private static ComplexList generateComplexList(int)`

aherbert (changes requested):
This looks OK. Just a few minor changes to address. Thanks

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -223,14 +222,87 @@ void testReplaceAllComplexScalarFunction() {
...
     /**
-     * Generates a list of random complex numbers of the given size.
+     * Generates a ComplexList of random complex numbers of the given size.
      * @param size number of complex numbers in the list.
-     * @return the list of random complex numbers.
+     * @return the ComplexList of random complex numbers.
      */
-    private static List<Complex> generateList(int size) {
-        return ThreadLocalRandom.current().doubles(size, -Math.PI, Math.PI)
+    private static ComplexList generateList(int size) {
```
aherbert, line 300: There are use cases for `generateList` and `generateComplexList`. The generateList would function as before and return a List implementation from the JDK. This can be assumed to work correctly. The generateComplexList would call generateList and then addAll to a new ComplexList instance. This should be asserted to be equal to the original List. This method can be used for convenience to populate a ComplexList for testing the accessor methods (get/set).

One case where you may wish to have a List that is not a ComplexList is in testing addAll. You also require a non ComplexList for any list compared for equality with ComplexList: you cannot have a list equals comparison to itself as there may be the same error in both implementations being compared.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -138,48 +139,122 @@ private void rangeCheckForInsert(int index) {
...
-     *
+     * <p>
      * {@inheritDoc}
+     *
      * @return the complex number.
      */
     @Override
     public Complex get(int index) {
         rangeCheck(index);
-        final int i = index << 1;
```
aherbert, line 148: The precomputed value i will be reused so it can be left as before, please revert.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -138,48 +139,122 @@ private void rangeCheckForInsert(int index) {
...
      * {@inheritDoc}
      *
-     * @param element Complex element to be set.
+     * @param number Complex number to be set.
      */
     @Override
-    public Complex set(int index, Complex element) {
+    public Complex set(int index, Complex number) {
         rangeCheck(index);
-        final int i = index << 1;
```
aherbert, line 174: The precomputed value i will be reused so it can be left as before, please revert.

In addition the call to setNoRangeCheck can be removed. Since you already have the value i this will avoid calling it again.

aherbert (changes requested):
I think you have removed too many uses of `i = index << 1`. There are some where this is useful.

Also the test class should have generateList and generateComplexList. The later to be used only when the list will not be compared to another list using assertEquals, e.g. for tests for accessor methods.

PR #124, opened 2022-09-08
Numbers 186.complex list interleaved

sumanth-rajkumar:
This PR refactors ComplexList to be abstract and gives a interleaved implementation as a private sub-class. This allows for multiple implementations of data storage. 

Summary of changes: 

ComplexList - made abstract with factory methods implemented in interleaved implementation.
ComplexInterleavedList - private sub-class with interleaved double array storage.
ComplexListTest - updates to calling ComplexList using ComplexInterleavedList.

### DIFF
--- a/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
+++ b/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
@@ -264,14 +264,14 @@ public abstract class ComplexList extends AbstractList<Complex> {
      * and permits all complex numbers. In addition to implementing the List interface,
      * this class provides methods to manipulate the size of the array that is used internally to store the list.
      *
-     * <p>Each ComplexList instance has a capacity. The capacity is half the size of the double array used to store the complex numbers
-     * in the list. As complex numbers are added to an ComplexList, its capacity grows automatically.
+     * <p>Each ComplexInterleavedList instance has a capacity. The capacity is half the size of the double array used to store the complex numbers
+     * in the list. As complex numbers are added to an ComplexInterleavedList, its capacity grows automatically.
      * The complex number is stored using an interleaved format and so the maximum number of complex numbers that may be added is
      * approximately 2<sup>30</sup>. This is half the maximum capacity of java.util.ArrayList.
      * The memory usage is more efficient than using a List of Complex objects as the underlying numbers are not stored
      * using instances of Complex.</p>
      *
-     * <p>An application can increase the capacity of an ComplexList instance before adding a large number of complex numbers
+     * <p>An application can increase the capacity of an ComplexInterleavedList instance before adding a large number of complex numbers
      * using the ensureCapacity operation. This may reduce the amount of incremental reallocation.</p>
      *
      * <p>This list does not support {@code null} Complex objects.
@@ -388,7 +388,7 @@ public abstract class ComplexList extends AbstractList<Complex> {
         }
 
         /**
-         * Increases the capacity of this ComplexList instance, if necessary, to ensure that it can hold at
+         * Increases the capacity of this ComplexInterleavedList instance, if necessary, to ensure that it can hold at
          * least the amount of complex numbers specified by the minimum capacity argument.
          *
          * @param minCapacity Desired minimum capacity.
@@ -420,7 +420,7 @@ public abstract class ComplexList extends AbstractList<Complex> {
         }
 
         /**
-         * Increases the capacity of this ComplexList instance, if necessary, to ensure that it can hold at
+         * Increases the capacity of this ComplexInterleavedList instance, if necessary, to ensure that it can hold at
          * least an additional amount of complex numbers specified by the capacity argument.
          *
          * @param capacity Desired capacity.
--- a/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
+++ b/commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
@@ -29,6 +29,11 @@ import java.util.Objects;
 
 /**
  * This is an abstract class that implements all the common data and operations for a ComplexList.
+ *
+ * <p>The memory usage is more efficient than using a List of Complex objects as the underlying numbers are not stored
+ * using instances of Complex.</p>
+ *
+ * <p>This list does not support {@code null} Complex objects.
  */
 public abstract class ComplexList extends AbstractList<Complex> {
 
@@ -50,6 +55,13 @@ public abstract class ComplexList extends AbstractList<Complex> {
 
     /** Size of complex numbers in the list. */
     protected int size;
+
+    /**
+     * Constructor to prevent extension of this class outside inner clases.
+     */
+    private ComplexList() {
+
+    }
 
     /** {@inheritDoc} */
     @Override
@@ -277,6 +289,7 @@ public abstract class ComplexList extends AbstractList<Complex> {
      * <p>This list does not support {@code null} Complex objects.
      */
     private static class ComplexInterleavedList extends ComplexList {
+
         /**
          * Storage for the complex numbers.
          * Data is stored in an interleaved format using (real, imaginary) pairs.
@@ -311,7 +324,7 @@ public abstract class ComplexList extends AbstractList<Complex> {
          * @throws IllegalArgumentException if the specified double array doesn't have an even amount of doubles.
          */
         ComplexInterleavedList(double[] fromArray) {
-            if (fromArray.length % 2 != 0) {
+            if ((fromArray.length & 1) != 0) {
                 throw new IllegalArgumentException("Length of array has to be even");
             }
             realAndImagParts = fromArray;
@@ -451,11 +464,13 @@ public abstract class ComplexList extends AbstractList<Complex> {
             if (size == e.length >>> 1) {
                 e = ensureCapacityInternal(size + 1);
             }
+            final double real = number.real();
+            final double imaginary = number.imag();
             final int i = index << 1;
             final int s = size << 1;
             System.arraycopy(e, i, e, i + 2, s - i);
-            e[i] = number.real();
-            e[i + 1] = number.imag();
+            e[i] = real;
+            e[i + 1] = imaginary;
             size++;
         }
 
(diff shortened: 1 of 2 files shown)


commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -28,23 +28,9 @@
...
- * using instances of Complex.</p>
- *
- * <p>An application can increase the capacity of an ComplexList instance before adding a large number of complex numbers
- * using the ensureCapacity operation. This may reduce the amount of incremental reallocation.</p>
- *
- * <p>This list does not support {@code null} Complex objects.
+ * This is an abstract class that implements all the common data and operations for a ComplexList.
  */
-public class ComplexList extends AbstractList<Complex> {
+public abstract class ComplexList extends AbstractList<Complex> {
```
aherbert, line 33: Prevent extension of this class outside the inner classes by adding a private constructor. It may have to be package-private to pass the build checks (e.g. PMD).

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -28,23 +28,9 @@
 import java.util.Objects;
 
 /**
- * Resizable-double array implementation of the List interface. Implements all optional list operations,
```
aherbert, line 31: Some of this javadoc still applies. The list provides efficient storage of complex numbers. It does not accept null complex objects.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -395,46 +229,322 @@ private String outOfBoundsMsg(int index) {
...
+        }
+
+        /**
+         * Constructs an interleaved list using the specified double array.
+         *
+         * @param fromArray Specified backing double array.
+         * @throws IllegalArgumentException if the specified double array doesn't have an even amount of doubles.
+         */
+        ComplexInterleavedList(double[] fromArray) {
+            if (fromArray.length % 2 != 0) {
```
aherbert, line 314: if ((fromArray.length & 1) != 0)

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -37,6 +37,17 @@ public class ComplexListTest {
 
     private static final int MAX_CAPACITY = (Integer.MAX_VALUE - 9) / 2;
 
+    @Test
+    void testFromArray() {
+        double[] fromArray1 = ThreadLocalRandom.current().doubles(6, -Math.PI, Math.PI).toArray();
```
aherbert, line 42:
```
int size = 3;
... doubles(size * 2, ...

for (... ; i < size; ... 
```

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -105,13 +116,31 @@ void testAddAndAddAll() {
...
 
+    @Test
+    void testSetAddAndAddAllNullPointerException() {
+        ComplexList list1 = generateComplexList(3);
+        Assertions.assertThrows(NullPointerException.class, () -> list1.add(null));
+        Assertions.assertThrows(NullPointerException.class, () -> list1.add(0, null));
+        Assertions.assertThrows(NullPointerException.class, () -> list1.set(1, null));
+
+        List<Complex> list2 = generateList(3);
+        for (int i = 0; i < list2.size(); i++) {
```
aherbert, line 134: This loop is setting all elements to null. Just set one of them:
`list2.set(1, null)`

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -105,13 +116,31 @@ void testAddAndAddAll() {
...
+            Complex val = list2.get(i);
+            if (val != null) {
+                list2.set(i, null);
+            }
+        }
+        Assertions.assertThrows(NullPointerException.class, () -> list1.addAll(list2));
```
aherbert, line 140: Check that list1 has the same size after the NPE is thrown, i.e. it is unchanged.

You could create a copy of the list for reference and use assertEquals to compare the before and after.

aherbert (changes requested):
This looks good. Just a few small changes required.

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -311,7 +324,7 @@ private static class ComplexInterleavedList extends ComplexList {
          * @throws IllegalArgumentException if the specified double array doesn't have an even amount of doubles.
          */
         ComplexInterleavedList(double[] fromArray) {
-            if (fromArray.length % 2 != 0) {
+            if ((fromArray.length & 1) != 0) {
                 throw new IllegalArgumentException("Length of array has to be even");
             }
             realAndImagParts = fromArray;
```
aherbert, line 330: Here the data is not defensively cloned. This is fine but the behaviour should be documented. Any public method should indicate that the argument array is used in-place; any external modifications to the array will be reflected in this List until a structural modification is made to the data storage, for example a resize.

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -39,9 +39,10 @@ public class ComplexListTest {
 
     @Test
     void testFromArray() {
-        double[] fromArray1 = ThreadLocalRandom.current().doubles(6, -Math.PI, Math.PI).toArray();
+        int size = 3;
+        double[] fromArray1 = ThreadLocalRandom.current().doubles(size * 2, -Math.PI, Math.PI).toArray();
         ComplexList list = ComplexList.from(fromArray1);
-        for (int i = 0; i < fromArray1.length >> 1; i++) {
+        for (int i = 0; i < size >> 1; i++) {
```
aherbert, line 45: This should iterate over `size`, not `size >> 1`

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -130,14 +131,12 @@ void testSetAddAndAddAllNullPointerException() {
...
         List<Complex> list2 = generateList(3);
-        for (int i = 0; i < list2.size(); i++) {
-            Complex val = list2.get(i);
-            if (val != null) {
-                list2.set(i, null);
-            }
-        }
+        list2.set(1, null);
         Assertions.assertThrows(NullPointerException.class, () -> list1.addAll(list2));
+        Assertions.assertEquals(copy.size(), list1.size());
```
aherbert, line 139: You should also check `addAll(int, Collection)`. The implementation, if done incorrectly, could move all elements to make space, then copy the new elements in and throw a NPE. This will leave the list in a broken state.

It may be better to check the contents of the list are the same using `Assertions.assertEquals(copy, list1);`

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -51,6 +56,13 @@ public abstract class ComplexList extends AbstractList<Complex> {
     /** Size of complex numbers in the list. */
     protected int size;
 
+    /**
+     * Constructor to prevent extension of this class outside inner clases.
+     */
+    private ComplexList() {
+
```
aherbert, line 63: Comment empty line: `// Intentionally empty`

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -130,14 +131,12 @@ void testSetAddAndAddAllNullPointerException() {
         Assertions.assertThrows(NullPointerException.class, () -> list1.add(0, null));
         Assertions.assertThrows(NullPointerException.class, () -> list1.set(1, null));
 
+        ComplexList copy = ComplexList.interleaved();
```
aherbert, line 134: Create this copy at the start of the method. When used at the end of the method this will also check the add methods do not result in modifications of the list.

aherbert (changes requested):

commons-numbers-complex-arrays/src/test/java/org/apache/commons/numbers/complex/arrays/ComplexListTest.java
```diff
@@ -126,18 +126,18 @@ void testAddAndAddAll() {
...
         Assertions.assertThrows(NullPointerException.class, () -> list1.set(1, null));
 
-        ComplexList copy = ComplexList.interleaved();
-        copy.addAll(list1);
         List<Complex> list2 = generateList(3);
         list2.set(1, null);
         Assertions.assertThrows(NullPointerException.class, () -> list1.addAll(list2));
-        Assertions.assertEquals(copy.size(), list1.size());
         Assertions.assertThrows(NullPointerException.class, () -> list1.addAll(0, list2));
+        copy.addAll(list1);
```
aherbert, line 139: copy.addAll(list1) should be **before** all the operations on list1. Otherwise you are not comparing before (the copy) and after (list1).

commons-numbers-complex-arrays/src/main/java/org/apache/commons/numbers/complex/arrays/ComplexList.java
```diff
@@ -60,7 +60,7 @@ public abstract class ComplexList extends AbstractList<Complex> {
      * Constructor to prevent extension of this class outside inner clases.
      */
     private ComplexList() {
-
+        //Intentional empty
```
aherbert, line 63: `// Intentionally empty`

aherbert (changes requested):

PR #128, opened 2023-03-09
Updated the checkstyle supressions file

kitswas:
Added missing regexes for paths to support both Windows and *nix.

On Windows, the path separator is `\` not `/`.
As a result, the build fails due to Checkstyle violations.


kitswas:
@aherbert Is this alright?

aherbert:
Looks good. Please feedback any other issues you have with building on Windows.

PR #130, opened 2023-07-13
[NUMB-194]: method that determines if 2 integers are coprime

orionlibs:
[NUMBER-194](https://issues.apache.org/jira/browse/NUMBERS-194). Overloaded method that determines if 2 integers are relatively prime or coprime. Included tests

### DIFF
--- a/commons-numbers-primes/pom.xml
+++ b/commons-numbers-primes/pom.xml
@@ -40,5 +40,11 @@
     <numbers.parent.dir>${basedir}/..</numbers.parent.dir>
     <numbers.jira.component>primes</numbers.jira.component>
   </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-numbers-core</artifactId>
+    </dependency>
+  </dependencies>
 
 </project>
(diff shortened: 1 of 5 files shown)


aherbert:
These are simply pass through methods for `ArithmeticUtils.gcd` where gcd == 1.

The new methods do not include a link to `gcd` and also lose all the javadoc on the special cases for that method, including exception cases. Adding these methods brings in the dependency on the core module and bloats the primes module. I believe it better to add javadoc to the Primes class to sign post users to ArithmeticUtils, e.g. in the class javadoc:

```
 * <p>Additional methods for integer arithmetic can be found in the Numbers core module in
 * {@code o.a.c.numbers.core.ArithmeticUtils}. This includes a method to obtain the
 * greatest common divisor (GCD) of two integers which can be used to determine if two numbers
 * are relatively prime.
```

Thus users can bring in that dependency if they require.

orionlibs:
> I was stating that this entire PR should be dropped and replaced with some better documentation in the Primes class header to indicate that the functionality to detect relatively prime numbers can be found in ArithmeticUtils.

@aherbert not every dev knows that checking for coprimes requires a GCD calculation, but they do know what coprimes require a primes utility. So, they will use the Primes class to see what kinds of methods it has. It will not occur to them to go to the ArithmeticUtils to check specifically the GCD method JavaDoc. I can add your suggested JavaDoc to the GCD methods, of course, but.............

PR #133, opened 2023-07-13
NUMBERS-194: GCD JavaDoc mention coprimes check

orionlibs:
[NUMB-194](https://issues.apache.org/jira/browse/NUMBERS-194) Changed JavaDoc of the GCD methods to mention that they can be used to determine if 2 numbers are relatively prime or coprime

### DIFF
--- a/commons-numbers-core/src/main/java/org/apache/commons/numbers/core/ArithmeticUtils.java
+++ b/commons-numbers-core/src/main/java/org/apache/commons/numbers/core/ArithmeticUtils.java
@@ -58,9 +58,9 @@ public final class ArithmeticUtils {
      *   for the special cases above.</li>
      *  <li>The invocation {@code gcd(0, 0)} is the only one which returns
      *   {@code 0}.</li>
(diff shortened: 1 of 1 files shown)


commons-numbers-core/src/main/java/org/apache/commons/numbers/core/ArithmeticUtils.java
```diff
@@ -58,6 +58,8 @@ private ArithmeticUtils() {
      *   for the special cases above.</li>
      *  <li>The invocation {@code gcd(0, 0)} is the only one which returns
      *   {@code 0}.</li>
+     *   <li>gcd can be used to determine if two numbers are relatively prime.
```
aherbert, line 61: This has been added as a special case. It is not a special case, it is a usage example. It should be moved to a new paragraph, e.g.
```
<p>Two numbers are relatively prime, or coprime, if their gcd is 1.
```

orionlibs: @aherbert fixed JavaDoc

PR #136, opened 2023-07-22
NUMBERS-200: Don't honour compensation if it's NaN in Sum#add

ani5rudh:


commons-numbers-core/src/main/java/org/apache/commons/numbers/core/Sum.java
```diff
@@ -202,7 +202,11 @@ public Sum subtract(final Sum other) {
      * @return this instance.
      */
     private Sum add(double s, double c) {
-        return add(s).add(c);
+        add(s);
+        if (!Double.isNaN(c)) {
```
aherbert, line 206: I think this can use `isFinite(c)`. I don't think the compensation can be infinite if the regular sum is still finite, so we do not have to sum infinite compensation.

However, the NaN check should be faster so the two variants are the same. Please add a comment as to why the NaN check is required, e.g. infinite terms will generate a nan compensation. Do not destroy the regular IEEE754 sum with spurious nan.

aherbert (changes requested):
This looks good. Just add a comment as to why we have a NaN check and then you can rebase and squash to a single commit.

